PR #8, opened 2023-04-13
first commit

LibenHailu:
- Since we have been instructed to vacate the dormitory there are somethings that are not done yet
- no husky rules
- no commit lint
- no stylelint since we use tailwind
- no test libs
- no redux toolkit
- no pages
- creating other folder is not done
- after the comments we can replicate them


YiseBoge (changes requested):
I can't let just the starter project be merged right now so I recommend finish up the needed setup and makeing the three team projects before merging

PR #12, opened 2023-04-18
implement navbar

Segnicho:


Web/starter_project_vanguard/components/example/Button.tsx
YiseBoge, line 1: I don't understand why the need to make changes to the example button
if you intend to use it i would recommend making another button or renaming it appropriately bacause example is just example i think
but if you are just experimenting with it, better not to track that on git

Web/starter_project_vanguard/styles/globals.css
```diff
@@ -17,11 +18,11 @@
 }
 
 body {
-  color: rgb(var(--foreground-rgb));
+  /* color: rgb(var(--foreground-rgb));
```
YiseBoge, line 21: better not to leave in comments in the final code
if you want to remove it then remove it 
we can always recover the code from the git history

Web/starter_project_vanguard/types/NavTypes.ts
```diff
@@ -0,0 +1,4 @@
+export interface NavType {
```
YiseBoge, line 1: might be better to name this something else 
like NavItem or sth because navtype implies some type of navigation (you can feel free to use similar names to what is used in the UI or tailwind or sth )

Web/starter_project_vanguard/components/common/Navbar.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Button } from '../example/Button'
+
+const montserrat = Montserrat({ subsets: ['latin'] })
+
+const Navbar = () => {
+  const router = useRouter()
+
+  const currSegment = router.pathname.split('/')[1] || ''
+
+  console.log(currSegment)
```
YiseBoge, line 18: no need for console.log on the final code on a PR

Web/starter_project_vanguard/components/common/Navbar.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <div className={montserrat.className}>
+      <div className="flex justify-between">
+        <Link href="/">
+          <Image src={logo} className="" alt="A2SV Logo" />
```
YiseBoge, line 24: [nit] is there need for a className that has no value in it?

YiseBoge (commented):
Nice work but I have laft some comments
additionally i would love it if you could leave some screenshots to review with descriptions

Web/starter_project_vanguard/pages/index.tsx
```diff
@@ -1,123 +1,14 @@
...
 import { Inter } from 'next/font/google'
+import Navbar from '@/components/common/Navbar'
 
 const inter = Inter({ subsets: ['latin'] })
 
 export default function Home() {
   return (
-    <main className="flex min-h-screen flex-col items-center justify-between p-24">
+    <main className="mx-10 my-5">
+      <Navbar/>
```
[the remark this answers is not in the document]
Segnicho: yes, you are right, it should be. in the experimental app folder structure, there is Layout.(tsx,..) in which we can put reusables like header and footer. in this case, i am not sure how. I'll look into it.

Web/starter_project_vanguard/components/common/Navbar.tsx
Segnicho: the navbar item name's font is Montserrat, which is why I used it here.

Segnicho:
Screenshot 2023-04-18 7 26 11 PM
I removed the Login tab as we don't have a page. and I tried to follow the DRY principle by not duplicating styling for the items.

Web/starter_project_vanguard/components/common/Navbar.tsx
YiseBoge: I am not talking about the font, I mean why do we have a className on `<Image src={logo} className="" alt="A2SV Logo" />`

Web/starter_project_vanguard/components/example/Button.tsx
```diff
@@ -33,7 +35,7 @@ export const Button = ({
       {...props}
       className={clsx(
         shape === 'square' ? 'rounded-md' : 'rounded-full',
-        'bg-blue-700 text-white p-8',
+        `bg-blue-700 text-white p-8`,
```
YiseBoge, line 38: [nit] why the need to change the quotation mark here ? 
if you are not using dynamic content in it I would recommend sticking to `''`

YiseBoge: and I also don't understand why the addition of spaces above (might be fine but I don't understand the purpose)

Web/starter_project_vanguard/pages/index.tsx
```diff
@@ -1,124 +1,9 @@
...
-          <p
-            className={`${inter.className} m-0 max-w-[30ch] text-sm opacity-50`}
-          >
-            Instantly deploy your Next.js site to a shareable URL with Vercel.
-          </p>
-        </a>
-      </div>
-    </main>
+    <div>
+      <Navbar />
```
YiseBoge, line 6: do we need to have the navbar here? or did we agree to have it inside a layout file?

YiseBoge: Actually I also don't think it is a good idea to remove all the index file content if you are not the one assigned to it (if you are to implement the new landing page yourself, that fine but otherwise you would just be creating conflict with people that you didn't need to)

Web/starter_project_vanguard/tailwind.config.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
         danger: '#b91c1c',
         'primary-text': '#353535',
         'secondary-text': '#717171',
+        'navColor':'#565656'
```
YiseBoge, line 17: should we not have the color be a primary color ?
what part would be different in the navColor from the primary color?
as in what color in the nav, should only be in the nav?

YiseBoge: [nit] also let us stick to `kebab-case` as it is used with the others instead of `camelCase`

Web/starter_project_vanguard/types/NavItem.ts
YiseBoge, line 1: I would recommend renaming this file to `nav.ts` or `layout.ts` to include any nav related or layout related types (the file name would be based on the feature)

and let us stick to kebab-case with the file naming here too

Web/starter_project_vanguard/utils/NavItems.ts
YiseBoge, line 1: [nit] let's use the kebab-case for this file too

I also think utils should be reserved for util functions, not mock data.

I have seen someone use the store folder to have mock items to load the data from maybe it is worth googling what is the best place for it

YiseBoge: personally I would recommend keeping this data inside the NavBar component itself

if you think about it, anyone that wants to edit this would go straight to that component and it is fine to have that data there

YiseBoge (commented):
I have left a few more comments

Web/starter_project_vanguard/types/NavItem.ts
Segnicho: recommendation accepted:)

Web/starter_project_vanguard/tailwind.config.js
Segnicho: the color on figma is not the same as what is defined primarily here..

Web/starter_project_vanguard/pages/index.tsx
Segnicho: I deleted the dummy starter file so that I can test the navbar from here. Okay then I will replace it with the dummy.

Web/starter_project_vanguard/components/example/Button.tsx
YiseBoge: [very nit] can you remove the extra space you added here if it is not useful?
It might create confusion when seeing changes on the file but there is nothing meaningful done

Web/starter_project_vanguard/components/layout/index.tsx
YiseBoge, line 1: I don't believe this file belongs in this folder 
I think you mean to make it a layout file that is shared between many pages

if so I would have it in a layout folder that is outside of components (this is not a component that you are including in pages, you use it differently I believe)

and I would also rename it to something like `default.tsx` or sth instead of `index.tsx`, because it is not just for the index page

Web/starter_project_vanguard/components/layout/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { ReactNode, Suspense } from 'react'
+import Navbar from './navbar'
+
+const RootLayout = ({ children }: { children: ReactNode }) => {
+  return (
+    <div className='px-20 py-8'>
```
YiseBoge, line 7: idk if it is a good idea to add padding at the layout level 
this just means the whole page will have a padding and you won't have the flexibility to decide such things on the page level

eski check what it looks like with the padding because I don't think even the nabber would look fine (not sure about this)

Web/starter_project_vanguard/components/layout/navbar.tsx
```diff
@@ -0,0 +1,51 @@
...
+const Navbar = () => {
+  const router = useRouter()
+  const currSegment = router.pathname.split('/')[1] || ''
+
+  return (
+    <div className="sticky top-0 bg-white bg-opacity-100 flex justify-between font-montserrat">
+      <Link href="/">
+        <Image src={logo} alt="A2SV Logo" />
+      </Link>
+      <div className="flex gap-4">
```
YiseBoge, line 26: this seems fine but can you check if tailwind has a tab that you can use for this list of links?
I think the animations and look of it might end up better if it exists

Web/starter_project_vanguard/pages/_app.tsx
```diff
@@ -1,6 +1,8 @@
 import '@/styles/globals.css'
 import type { AppProps } from 'next/app'
 
+import Layout from '@/components/layout'
+
 export default function App({ Component, pageProps }: AppProps) {
-  return <Component {...pageProps} />
+  return <Layout><Component {...pageProps} /> </Layout>
```
YiseBoge, line 7: is there a way to decide which layout to use for each page? (incase we end up making multiple layouts)
like for index page I could want a layout that has a footer in it, for login I may want one that has only navbar or sth

I just want you to check if there is an existing way to deal with layouts in NextJS but for now this should work fine too

Web/starter_project_vanguard/pages/index.tsx
```diff
@@ -121,4 +121,4 @@ export default function Home() {
       </div>
     </main>
   )
-}
+}
```
YiseBoge, line 124: [very nit] same comment as before, this is not a problem but it might be confusing to see changes on a file but have that only be some new line so if possible I would remove this small change too

Web/starter_project_vanguard/tailwind.config.js
```diff
@@ -14,12 +14,19 @@ module.exports = {
...
         'secondary-text': '#717171',
+        'tertiary-text': '#565656',
+        
       },
       backgroundImage: {
         'gradient-radial': 'radial-gradient(var(--tw-gradient-stops))',
         'gradient-conic':
           'conic-gradient(from 180deg at 50% 50%, var(--tw-gradient-stops))',
       },
+      fontFamily: {
```
YiseBoge, line 25: nice
I have seen someone else that did this as well so if there exists conflict when merging just be careful
otherwise I like it

YiseBoge (approved):
Overall I like this, I have left some minor comments that you can address before merging if possible

Web/starter_project_vanguard/components/layout/index.tsx
Segnicho: so, flexibility is good, but it costs us consistency. I can give padding for the navbar only, and if others give different padding for their pages, they will have different paddings and it looks weird, I thought. but ok I will give the padding for the nav only.

Web/starter_project_vanguard/pages/_app.tsx
Segnicho: Yes, there is. but it isn't necessary for this project as all the pages have the same footer and header.

Web/starter_project_vanguard/tailwind.config.js
Segnicho: Okay.

Web/starter_project_vanguard/components/layout/navbar.tsx
Segnicho: Screenshot 2023-04-22 3 33 06 PM
when active the tab's color and border change, if that is what you are asking.

YiseBoge: yeah that is what I mean but I can guarantee you they probably implemented it better
It is fine to merge with this for now but I recommend you switch to an existing tab even if it is in another PR

https://tailwindui.com/components/application-ui/navigation/tabs
Screenshot 2023-04-22 at 10 30 34 PM
example from their website (which one do you think looks better?)

I recommend you make a habit of using things that are already built for you if it exists

YiseBoge (approved):
This looks fine to merge for this PR for me

PR #13, opened 2023-04-20
feat(web-barbies) : Add blog card component

TsehaynehGetaneh:
Adding a new blog card feature
- Created a new "blog card" component that displays the blog image, title of the blog, author profile, author name, blog date of published, content of blog, selected tags, likes button, and read more button for each blog post.

please, review the implementation I did.

blogCard-likes
blogCard-pending


Web/starter_project_barbies/types/blog/Blog.ts
YiseBoge, line 1: [nit] does this need to be it's own file within a folder ?

think of what other blog related types there are

if they can be included inside this file, then i would recommend not having a blog folder within types folder and just having a blog.ts that holds blog ralated types (also no need for capital letter naming of the file here, let's stick to kebab case for all things but components)

Web/starter_project_barbies/utils/dateFormatter.ts
```diff
@@ -0,0 +1,4 @@
+export function formatDate(date : Date) {
```
YiseBoge, line 1: this is nice but it only formats in a specific format right ?
So if someone else were to use date formatting in a different format, they would use a different function

now the naming implies that this is the formatDate function and no other exists 
so either make it so that this does all kinds of formatting and you take in the format that people want to use, or make sure the function name implies what kind of format it does

YiseBoge: personally i would rather use prebuilt formatting libraries that take in the format (like moment js - nor sure if it is too big a library tho) instead of having my own util function

YiseBoge (approved):
Looks pretty good but i have left some comments that i trust you'll fix before merging

TsehaynehGetaneh:
All Issues are fixed now.

Web/starter_project_barbies/components/blog/BlogCard.tsx
```diff
@@ -0,0 +1,72 @@
+import Image from 'next/image';
+import Link from 'next/link';
+import { Blog } from '@/types/blog';
+import LikeButton from './LikeButton';
+import moment from "moment";
+
+
+// date formatter function using moment.js
+const formatDate = (dateString: string): string => {
+  const date = moment(dateString);
```
YiseBoge, line 10: nice

Web/starter_project_barbies/types/blog.ts
```diff
@@ -0,0 +1,16 @@
+// Each blog has these properties
+export type Blog = {
+    userID : string;
+    blogID: string;
+    title: string;
+    content: string;
+    tags: string[];
+    date: string;
+    author: {
```
YiseBoge, line 9: for now this is fine but when data is loaded this should be modified where userId is either in the autor part, or you only have userId and you have to load the author separately (based on how the backend would work)

but nice for now

YiseBoge (approved):
Looks Good. Great Job

YiseBoge:
> All Issues are fixed now.

usually resolve the comments when you fix that issue so that it is clearer to see what is solved and what is not

PR #14, opened 2023-04-21
feat(Web-Iconics): create a common blogCard component

blenassefa2:
On this branch a new common component called BlogCard is created in "/starter_project_iconics/components/common" in addition two interfaces 'Blog' and 'User' are created in "/starter_project_iconics/types"  and other css file and mock data in their respective folders.

On the figma file the blog card component is shared between two pages as follows
page1: 
image-20230419-091457

page2:
image-20230419-091400

In this branch the created card uses props to be reused in both pages and looks as follows:

image


Web/starter_project_iconics/components/common/blogCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+} from 'react-icons/ai'
+
+interface Props {
+  card: Blog
+  feature: string
+}
+
+const BlogCard: React.FC<Props> = ({ card, feature }) => {
+  const handleReadMore = () => {
+    console.log('Read More button clicked')
```
YiseBoge, line 19: better not to have console log in a mergeable code

Web/starter_project_iconics/components/common/blogCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+  const statusIcons: ReactElement[] = [
+    <AiFillCheckCircle key={1} size={20} className={styles.statIcon} />,
+    <AiFillWarning key={2} size={20} className={styles.statIcon} />,
+    <AiFillClockCircle key={3} size={20} className={styles.statIcon} />,
+  ]
+  const statusToNum: (stat: string) => number = (stat) => {
```
YiseBoge, line 26: [nit] This is a pretty good idea but it might be an even better approach to use a key-value pair instead of a list that you index to (you don't need to convert the strings "approved", "declined", [anything else] into numbers)

if for instance you had a object that had the strings as keys and the elements as values `{approved: <AiFillCheckCircle..., denied: <AiFillWarning..., none: <AiFillClockCircle...}`

then later on you can refer to it as `statusIcons[card.status]`
or you can even use `statusIcons[card.status] || statusIcons.none`, which means that it will always default to non if the status is not found as a key (meaning if it is not approved or denied)

yours works just fine, i am just suggesting because i feel it makes less code and it gets clearer which one each icon coresponds to when you define the list (usually using a list is not good in these cases because it might confuse a reader of the code as to why things are in that order. lists are for things you want to loop through if possible)

Web/starter_project_iconics/components/common/blogCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+    if (stat === 'approved') return 0
+    if (stat === 'denied') return 1
+    return 2
+  }
+  const formatNumber: (num: number) => string | number = (num: number) => {
```
YiseBoge, line 31: is there maybe a number formatting library you can use here ?
if possible that would be a nice idea

[nit] if you want to shorten this code you can use a similar method as above and define the letters "K", "M", "B" to a hashmap and index them by using some log math to get the index (just a suggestion for a coding challenge)

Web/starter_project_iconics/components/common/blogCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+                      style={{ display: 'inline-block' }}
+                    >
+                      {formatNumber(card.likes)} likes
+                    </span>
+                  </span>
+                )
+              } else {
+                return (
+                  <span style={{ display: 'flex', alignItems: 'center' }}>
+                    {
```
YiseBoge, line 105: [very nit] i am not sure what this brace is for surrounding the span

Web/starter_project_iconics/package.json
```diff
@@ -7,7 +7,7 @@
   },
   "scripts": {
     "dev": "next dev",
-    "build": "next build",
+    "build": "next build ",
```
YiseBoge, line 10: idk why we need a space after "build"

Web/starter_project_iconics/pages/profile/index.tsx
```diff
@@ -0,0 +1,35 @@
+import BlogCard from '@/components/common/blogCard'
+import { card1 } from '@/store/mockData/data'
+
+const index = () => {
+  const nums: number[] = [0, 1, 2, 3, 4]
```
YiseBoge, line 5: there might be a better way to loop 4 times other than creating a list of numbers
reading down, it looked to me like it had another purpose that i didn't see

at least I would recommend making a list of what kind of data would be populated in the card (like [{card: card1, feature: "likes"}]) and duplicating them to have four items 

at least that way I know what that list represents (and the naming of it should also explain that)

YiseBoge: or if the data is coming from mock anyways, it might be better to have that list of duplicates be created in the mocks just the way you would use it

Web/starter_project_iconics/pages/profile/index.tsx
```diff
@@ -0,0 +1,35 @@
+import BlogCard from '@/components/common/blogCard'
+import { card1 } from '@/store/mockData/data'
```
YiseBoge, line 2: might be a better idea to rename card1
idk what card1 means as opposed to card2? is there even card2?

Web/starter_project_iconics/pages/profile/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+        flexDirection: 'row',
+        justifyContent: 'flex-start',
+        flexWrap: 'wrap',
+        alignContent: 'flex-end',
+      }}
+    >
+      {nums?.map((value, index) => {
+        if (index % 2 === 0)
+          return (
+            <div style={{ marginLeft: '7px' }} key={value}>
```
YiseBoge, line 20: [nit] I think it might be a better idea to add the margin to the card itself (and the key as well)
i don't see a reason why we would need a wrapper component for it

Web/starter_project_iconics/pages/profile/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+        display: 'flex',
+        marginTop: '10px',
+        flexDirection: 'row',
+        justifyContent: 'flex-start',
+        flexWrap: 'wrap',
+        alignContent: 'flex-end',
+      }}
+    >
+      {nums?.map((value, index) => {
+        if (index % 2 === 0)
```
YiseBoge, line 18: if we have the list generated like i mentioned above, we wouldn't need an if condition to check the index. we can just pass the "feature" as it comes from the list (which i think is the only difference between the two)

Web/starter_project_iconics/store/mockData/data.tsx
YiseBoge, line 1: is the store folder the right place to hold mockData?

I am not sure it might be, but I had assumed the store folder is to save redux code. i just think it might be worth checking one more time

Web/starter_project_iconics/styles/blogCard.module.css
YiseBoge, line 1: if at all possible i would recommend using "scss" files for our styles (i think it should work when you import it without issue)

YiseBoge: I would also recommend trying to use tailwind classes as much as possible instead of custom CSS

I would minimize the CSS I would write as much as I can

cause this means the colors, fonts spaces... used here are not what other people are using and there will surely be inconsistencies

Web/starter_project_iconics/types/models.tsx
YiseBoge, line 1: it might be a good idea to name the file based on the feature it relates to rather than calling it "models"

like we could have user-related types in one user file, blog-related types in another, team-related types...

this might be fine for now but when other people's codes are included, it can get confusing

YiseBoge (commented):
Overall this is actually pretty great
i have left some comments tho (mostly minor)

YiseBoge:
i also wanted to remind that we are not using the [team] commit message syntax anymore
you can check out the git best practices doc (or just google "conventional commit") and we will be using that pattern for consistency

Adona12:
On the third picture you attached, it says approved on the bottom of the card. I don't think that's on the Figma desgn. The font weight of "read more" isn't consistent with the Figma design.

Adona12:
The date that's displayed on the card isn't also in the same format as the one on the Figma design.

YiseBoge (changes requested):
i have left some comments

Web/starter_project_iconics/components/common/blogCard.tsx
blenassefa2: I've changed it now

Web/starter_project_iconics/components/common/blogCard.tsx
blenassefa2: I agree and it was pretty easy to change to dictionary and now it is changed

Web/starter_project_iconics/components/common/blogCard.tsx
blenassefa2: I have found a way that is not so different but looks simpler and now it is changed

blenassefa2:
the commit got messed up so I will work on a different branch

PR #18, opened 2023-04-22
feat(web-vanguard): implement the Member card component

Yared04:
member


Adona12:
Can you attach a screenshot of what you implemented?

Yared04:
member

YiseBoge:
minor point: stick to kebab-casse with the branch naming (yared.member-page) for consistency

Web/starter_project_vanguard/components/team/Member.tsx
```diff
@@ -0,0 +1,36 @@
...
+import Link from "next/link"
+
+interface MemberProps {
+    name?: string,
+    picture?: string,
+    role?: string,
+    description?: string,
+    links?: string[]
+}
+const Member = ( {name = "Yared Tegegn", picture="https://media.licdn.com/dms/image/C4E03AQEVlrjCsy97_Q/profile-displayphoto-shrink_800_800/0/1661101696311?e=2147483647&v=beta&t=JK2LQ6ZoiQeWxoLnrPUwUiX5Bqi5mHBTIVKh4HK0cQk", role="Intern", description="Yared is a motivated software engineer looking to pursue a successful career in software development where he can help deliver software solutions to social problems. His skills include Web Development, Data Structures, Machine Learning, and DevOps. Natnael’s hobbies include coding, playing soccer, and watching movies.", links=["https://github.com/", "https://github.com/", "https://github.com/"]}: MemberProps) => {
```
YiseBoge, line 11: 1. can you use some syntax like React.FC or sth to define the component here (took me a while to recognize that it is a component)
2. please don't set the defaults here (i don't think the component needs defaults in the first place) - I recommend you pass your default member object from where you are rendering it multiple times if you have to
3. If possible try to find a better place to define the data and the Member type as it might be a type that could be useful in multiple components (or maybe not idk, but at least the data stored here within the code just creates confusion) - you might even use a JSON to store the mock data and see how you would render that
4. might be better to rename it as `MemberCard` as it better describes what it would look like (also remember to update the prop interface accordingly too)

Web/starter_project_vanguard/components/team/Member.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <div className="bg-white rounded-md drop-shadow-xl py-6 px-6 max-w-sm ">
+        <Image className="rounded-full m-auto text-center" width={120} height={120} src={picture} alt={name} />
+        <h2 className="font-bold mt-4 uppercase text-2xl text-center text-primary-text">{name}</h2>
+        <h5  className="font-medium uppercase text-center text-primary-text">{role}</h5>
+        <p className="text-center text-secondary-text my-3">{description}</p>
+        <hr className="border mt-7 mb-3" />
+        <div className="flex justify-around">
+          {
+            links.map((link: string, idx: number) => {
+              console.log(idx, link, socials[idx]);
```
YiseBoge, line 23: better not to have console log inside a mergable code (try not to commit it next time if possible)

Web/starter_project_vanguard/components/team/Member.tsx
```diff
@@ -0,0 +1,36 @@
...
+              return(
+              <Link key={idx} href={link} target="_blank">
+                <Image className="hover:" width={27} height={27} src={socials[idx]} alt={socials[idx]} />
```
YiseBoge, line 26: I don't like how you are using the socials here, i think the socials are accociated with each member so they should probably be defined together

i recommend defining the links with sth like:
```
...
links: [
  {icon: "facebook.svg", url: "faceboook.com..."},
  {icon: "gothub.svg", url: "github.com..."}
  ...
]
```
then loop through those as they are defined in your mock data

Web/starter_project_vanguard/components/team/Member.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const socials = ["facebook.svg", "linkedin.svg", "insta.svg"]
+  return (
+    <div className="bg-white rounded-md drop-shadow-xl py-6 px-6 max-w-sm ">
+        <Image className="rounded-full m-auto text-center" width={120} height={120} src={picture} alt={name} />
+        <h2 className="font-bold mt-4 uppercase text-2xl text-center text-primary-text">{name}</h2>
+        <h5  className="font-medium uppercase text-center text-primary-text">{role}</h5>
+        <p className="text-center text-secondary-text my-3">{description}</p>
```
YiseBoge, line 18: [nit] maybe try text-justify and see if that looks better

YiseBoge (changes requested):

YiseBoge (changes requested):

Web/starter_project_vanguard/types/teams.ts
```diff
@@ -0,0 +1,9 @@
+type SocialMedia = { url: string; social: string }
+
+export type Member = {
```
YiseBoge, line 3: [very-nit] we can rename this to `TeamMember` to stay consistent with other teams if we ever do another PR

Web/starter_project_vanguard/pages/_app.tsx
```diff
@@ -1,8 +1,12 @@
...
 import type { AppProps } from 'next/app'
 
-import RootLayout from '@/layout/default' 
+import RootLayout from '@/layout/default'
 
 export default function App({ Component, pageProps }: AppProps) {
-  return <RootLayout><Component {...pageProps} /> </RootLayout>
+  return (
+    <RootLayout>
+      <Component {...pageProps} />{' '}
```
YiseBoge, line 9: [nit] let's remove `{' '}` that is auto generated

Web/starter_project_vanguard/components/teams/Members.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import Member from './MemberCard'
+import members from '../../data/team-members.json'
```
YiseBoge, line 3: [nit] it might be a better idea to stick with the same naming as the file (teamMembers)
same goes for the component too (MemberCard)

Web/starter_project_vanguard/components/teams/MemberCard.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { IconContext } from 'react-icons/lib'
+
+interface MemberCardProps {
+  member: Member
+}
+const MemberCard: React.FC<MemberCardProps> = ({ member }) => {
+  let social_icon_map = new Map<string, JSX.Element>()
+  social_icon_map.set('facebook', <FaFacebook />)
+  social_icon_map.set('linkedin', <FaLinkedin />)
+  social_icon_map.set('instagram', <FaInstagram />)
```
YiseBoge, line 15: 1. let's stick to camelCase for variable
2. it might be cleaner to just build an object (key value pair)
```
const socialMediaIcons = {
  facebook: <FaFacebook />,
  linkedin: <FaLinkedin />
  instagram: <FaInstagram />
  default: [some default icon to use when it is not in here, could be FaWeb or sth]
}
```

and when using it you can do `socialMediaIcons[link.social] || socialMediaIcons.default` to make it use the default when undefined

YiseBoge: and for general knowledge, you don't need to use `let` as long as you are not reassigning the variable. it should stay as `const`

Web/starter_project_vanguard/components/teams/MemberCard.tsx
```diff
@@ -0,0 +1,51 @@
...
+      </h5>
+      <p className="text-center text-secondary-text my-3">
+        {member.description}
+      </p>
+      <hr className="border mt-7 mb-3" />
+      <div className="flex justify-around">
+        {member.links.map((link, idx: number) => {
+          return (
+            <Link key={idx} href={link.url} target="_blank">
+              <IconContext.Provider value={{ color: '#717171', size: '27px' }}>
```
YiseBoge, line 40: [nit] Let's use global color values and sizes (if they don't exist feel free to define them to tailwind)

YiseBoge (dismissed):
I have left some more nits but I'll approve and it might be a better idea to solve them as a FLUP (follow-up PR)

Adona12 (approved):
LGTM

PR #24, opened 2023-04-22
feat(web-iconics):  Implement header component 

Aklile-Yilma:
nav-bar-desktop
nav-bar-mobile


Henok-Matheas (approved):
LGTM. make the commit messages present tense tho.

YiseBoge:
I agree about the commit message (please check again the guideline documents or google conventional commits) - also use `feat` instead of `feature` for consistency

Also please attach screenshots and descriptions of what you implemented

Web/starter_project_iconics/components/layout/Header.tsx
YiseBoge, line 1: `Navbar` might be a better file name for consistency

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
+import Image from 'next/image'
+import { useState } from 'react'
+import { MdOutlineClose, MdViewHeadline } from 'react-icons/md'
+import logo from '../../public/a2sv-logo.png'
+
+function classNames(...classes: string[]) {
+  return classes.filter(Boolean).join(' ')
+}
+
+const Header = () => {
```
YiseBoge, line 10: can you use a syntax like React.FC (or whatever is the proper way for next) to define components?
It clarifies what is what

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
+import Image from 'next/image'
+import { useState } from 'react'
+import { MdOutlineClose, MdViewHeadline } from 'react-icons/md'
+import logo from '../../public/a2sv-logo.png'
```
YiseBoge, line 4: i feel like there should be another way to access the public folder than going back a couple of folders

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+function classNames(...classes: string[]) {
+  return classes.filter(Boolean).join(' ')
+}
+
+const Header = () => {
+  const [navigation, setNavigation] = useState(() => [
+    { id: 0, name: 'Home', to: '/', current: true },
```
YiseBoge, line 12: 1. idk why you need the id section as long as it is the same as the index
2. idk if tracking current this way is the best (i think current might be best tracked as a separate number - if you for instance want to set the current based on the url that is loaded, you can have a function that can do that on page load)
3. why do we have `/Success_story/success_stories` as the link? i don't understand why they have to be sub folders (also let us stay consistent with kebab-case for them too - not underscores and capitals)
4. i think there might be preexisting tabs that you can use that do this (check if they exist)

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+      name: 'Get Involved',
+      to: '/get_involved/get_involved',
+      current: false,
+    },
+  ])
+
+  const [showNav, setShowNav] = useState(false)
+
+  const handleNavClick = (e: React.FormEvent) => {
+    let id = parseInt((e.target as HTMLAnchorElement).id)
```
YiseBoge, line 33: [nit] Idk where id is being modified, so better to use const

YiseBoge: i think we don't need this too if we remove the id
you can just have them link to their pages and set the `current` based on the current route

Web/starter_project_iconics/components/layout/Header.tsx
YiseBoge, line 4: [nit] maybe better to call this a2svLogo for clarification

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+    )
+  }
+
+  return (
+    <div className="bg-white flex justify-between p-5">
+      <div className="w-28 lg:w-52">
+        <Image
+          src={logo}
+          alt="a2sv-logo"
+          width={200}
```
YiseBoge, line 52: i recommend only setting one (width or height) and letting the image scale on it's own (this would make it easier to make changes and won't distort or crop the image)
and i don't think we want cover for this logo as we want it to sit as it is

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+          height={50}
+          className="object-cover"
+        />
+      </div>
+
+      <div className="w-3/4 flex flex-col gap-y-4  md:ml-0 md:flex-row justify-between">
```
YiseBoge, line 58: [very-nit] `gap-y-4  md:ml-0` has an extra space in between

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+        )}
+
+        <nav
+          className={classNames(
+            showNav ? 'flex' : 'hidden',
+            'md:flex ml-auto md:ml-0 md:w-3/4'
+          )}
+        >
+          <ul className="flex flex-col gap-5 content-between md:flex-row md:w-full justify-between">
+            {navigation.map(({ id, name, to, current }) => (
```
YiseBoge, line 82: you can access the index here using `navigation.map((({ name, to, current }), index) => (...` instead of saving it with the data

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+              <li key={id}>
+                <a
+                  href={to}
+                  id={id.toString()}
+                  className={classNames(
+                    current
```
YiseBoge, line 88: if we make the change to save current as a string, the check here would have to change to `index === current`

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+                      ? 'text-blue-700 underline underline-offset-8 decoration-4'
+                      : 'no-underline text-gray-900',
+                    'font-semibold text-base lg:text-xl'
+                  )}
+                  onClick={(e: React.FormEvent) => handleNavClick(e)}
```
YiseBoge, line 93: can you not just pass `handleNavClick` for the onClick directly?

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <nav
+          className={classNames(
+            showNav ? 'flex' : 'hidden',
+            'md:flex ml-auto md:ml-0 md:w-3/4'
+          )}
+        >
+          <ul className="flex flex-col gap-5 content-between md:flex-row md:w-full justify-between">
+            {navigation.map(({ id, name, to, current }) => (
+              <li key={id}>
+                <a
```
YiseBoge, line 84: there might be a buttton component that can look like a link but acts like a button (personally i like the looks of those)

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+          src={logo}
+          alt="a2sv-logo"
+          width={200}
+          height={50}
+          className="object-cover"
+        />
+      </div>
+
+      <div className="w-3/4 flex flex-col gap-y-4  md:ml-0 md:flex-row justify-between">
+        {showNav ? (
```
YiseBoge, line 59: not sure i understand how the showNav works
I am hoping it will make better sense with the screenshots

Web/starter_project_iconics/components/layout/Header.tsx
```diff
@@ -0,0 +1,118 @@
...
+                  {name}
+                </a>
+              </li>
+            ))}
+          </ul>
+        </nav>
+
+        <div
+          className={classNames(
+            showNav ? 'flex' : 'hidden',
```
YiseBoge, line 104: still don't understand what is is meant to do

Web/starter_project_iconics/package-lock.json
YiseBoge, line 1: i am not sure why there is a change in the package-lock.json without any changes in the package.json 
maybe try running `npm install` to make sure things are proper

YiseBoge (changes requested):
Hey, it would be nice to get screenshots to understand some things but i have left comments on the code

Web/starter_project_iconics/components/layout/Header.tsx
Adona12: I think you should attach screenshot of both web and mobile view

Aklile-Yilma:
nav-bar-desktop
nav-bar-mobile

sam5235 (approved):

Web/starter_project_iconics/pages/_app.tsx
YiseBoge, line 1: let's create a layout file (this can be a different PR)

Web/starter_project_iconics/data/layout/navbar/nav-data.ts
YiseBoge, line 1: let's also move this back to the navbar (can be handled on the other PR)

Web/starter_project_iconics/components/layout/Navbar.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }, [asPath])
+
+  return (
+    <div className="bg-white flex justify-between p-5">
+      <div className="w-28 lg:w-52">
+        <Image src="/a2sv-logo.png" alt="a2sv-logo" width={200} height={50} />
+      </div>
+
+      <div className="w-3/4 absolute -right-0 mr-4 flex flex-col gap-y-4 md:ml-0 md:flex-row  md:static justify-between">
+        {showNav ? (
```
YiseBoge, line 37: it might be a good idea to add a comment to describe what this is for

Web/starter_project_iconics/components/layout/Navbar.tsx
```diff
@@ -0,0 +1,95 @@
...
+        </nav>
+
+        <div
+          className={classNames(
+            showNav ? 'flex' : 'hidden',
+            'ml-auto md:ml-0 md:flex gap-4 text-sm lg:text-lg'
+          )}
+        >
+          <button className="font-bold m-4 md:m-0">Login</button>
+          <button className="bg-blue-700 text-white font-bold px-3 rounded-2xl">
```
YiseBoge, line 86: let's use the global colors when they are ready

Web/starter_project_iconics/components/layout/Navbar.tsx
```diff
@@ -0,0 +1,95 @@
...
+            showNav ? 'flex' : 'hidden',
+            'md:flex ml-auto md:ml-0 md:w-3/4'
+          )}
+        >
+          <ul className="flex flex-col gap-5 content-between md:flex-row md:w-full justify-between">
+            {navigation.map(({ name, to, current }, index) => (
+              <li key={index}>
+                <Link
+                  href={to}
+                  id={index.toString()}
```
YiseBoge, line 64: i don't think this is necessary anymore

Web/starter_project_iconics/components/layout/Navbar.tsx
```diff
@@ -0,0 +1,95 @@
...
+                >
+                  {name}
+                </Link>
+              </li>
+            ))}
+          </ul>
+        </nav>
+
+        <div
+          className={classNames(
```
YiseBoge, line 80: one other approach 
```
{
  mlAuto,
  ...
  flex: showNav,
  hidden: !showNav
}
```

YiseBoge, line 80: one other approach 
```
{
  mlAuto,
  ...
  flex: showNav,
  hidden: !showNav
}
```

YiseBoge (approved):
This looks good
i have left some nits that you can handle as a FLUP

PR #25, opened 2023-04-22
feat(web-vanguard): implement the Member card component 

Yared04:
Changed the branch name.
#18


YiseBoge (changes requested):
check out all my comments on the previous PR
you did not need to create a new PR btw, you could have just renamed the branch there (if that is all you changed)
Personally i would recommend if we closed this and continued with the previous PR because i have left some comments there already

Yared04:
> check out all my comments on the previous PR you did not need to create a new PR btw, you could have just renamed the branch there (if that is all you changed) Personally i would recommend if we closed this and continued with the previous PR (if nothing else is different) because i have left some comments there already i also like the naming of the previous PR better

Yared04:
> check out all my comments on the previous PR you did not need to create a new PR btw, you could have just renamed the branch there (if that is all you changed) Personally i would recommend if we closed this and continued with the previous PR (if nothing else is different) because i have left some comments there already i also like the naming of the previous PR better

When renaming the branch, this branch gets deleted and a new one is created with the same content, so I had to open a new PR.
I have modified this pr based on the comments you gave on the previous pr.
Also you asked to change the name of the pr on your first comment

Yared04:
member

PR #27, opened 2023-04-22
feat(web-vanguard): Implement home page landing section

Rediet-Ferew:
In this branch the landing section of the web page has been implemented in @components/home/Landing_section

The resulting page is:
Home_landing


YiseBoge:
some general comments
1. The UI looks nice but let's check back to make sure it works when the navbar is added as well
2. please edit the PR name to follow the conventional commit rules as it will be the default commit message when doing squash and merge

Web/starter_project_vanguard/components/home/Landing_section/Buttons.tsx
YiseBoge, line 1: i don't think it is necessary to have a Landing_section folder within the home folder
one landing folder should be able to take care of most landing page related components in my opinion
and the naming might be good to call it landing as that is what we have been calling the page in the designs as well if i am not mistaken

YiseBoge: and let's remember to keep using kebab-case with file names

Web/starter_project_vanguard/components/home/Landing_section/Buttons.tsx
```diff
@@ -0,0 +1,33 @@
+
+import Link from 'next/link'
+export const Buttons = () => {
+  return (
+    <div className="flex w-full justify-start">
+      {/* get started button goes here */}
```
YiseBoge, line 6: [nit] i think it might be better not to say "goes here" it implies that there is something that is to be added in place of the comment

Web/starter_project_vanguard/components/home/Landing_section/Buttons.tsx
```diff
@@ -0,0 +1,33 @@
+
+import Link from 'next/link'
+export const Buttons = () => {
```
YiseBoge, line 3: it might be better to use a syntax like React.FC (or whatever is the best wat for nextJS) to show what is what clearly

Web/starter_project_vanguard/components/home/Landing_section/Buttons.tsx
YiseBoge: I also don't understand what the "Buttons" are so it might be better to use a more descriptive name for the component

Web/starter_project_vanguard/components/home/Landing_section/Landing.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <div className="mt-20 mb-10">
+      <div className="grid w-full grid-cols-2 gap-10 p-10 pl-4 mt-8 place-content-center place-items-center">
+        {/* left part goes here */}
+        <div className="w-full ml-4">
+          <div>
+            {/* heading goes here */}
+            <h1 className="text-4xl font-extrabold sm:mb-7">
+              Africa To
+              <span className="text-blue-600"> Silicon Valley</span>
```
YiseBoge, line 15: [nit] might be a good idea to reuse global colors as much as possible (idk maybe primary color could work for this for example)

Web/starter_project_vanguard/components/home/Landing_section/Landing.tsx
```diff
@@ -0,0 +1,46 @@
...
+            </h1>
+
+            {/* text goes here */}
+            <p className="mt-3 text-gray-700 whitespace-normal font-body md:text-lg md:font-semibold">
+              A2SV up-skills high-potential university students, connects them
+              with opportunities at top tech companies.
+            </p>
+
+            {/* buttons go here */}
```
YiseBoge, line 24: "buttons" does not describe what buttons these are supposed to be (especially for someone that has been reviewing different parts of the doc) so I would remind the naming improvement

Web/package-lock.json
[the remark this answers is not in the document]
YiseBoge: wait this is in the web folder 
it should not exist in the first place

Web/starter_project_vanguard/package-lock.json
YiseBoge, line 1: why does the package-lock.json file have changes when the package.json?
please do an npm install to make sure everything is well

Web/starter_project_vanguard/components/home/Landing_section/Landing.tsx
```diff
@@ -0,0 +1,46 @@
...
+            </div>
+          </div>
+        </div>
+
+        {/* right part goes here */}
+        <div className="relative w-full mt-20 aspect-square place-self-start">
+          {/* image element goes here */}
+          <Image
+            src="/images/landing_image.png"
+            alt="Picture of the author"
```
YiseBoge, line 36: is it a picture of the author?

Web/starter_project_vanguard/pages/_app.tsx
```diff
@@ -1,6 +1,5 @@
 import '@/styles/globals.css'
 import type { AppProps } from 'next/app'
-
```
YiseBoge, line 3: please let us avoid changes that have nothing to do with the feature (if it is something that needs to happen, it can be a different PR)

i would say bring back this change

Web/starter_project_vanguard/components/home/Landing_section/Landing.tsx
YiseBoge, line 1: I would rename this to hero section to stay consistent with others too and because we are calling the page landing page and we would not want to create confusion

Web/starter_project_vanguard/styles/globals.css
```diff
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css2?family=Poppins:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&display=swap');
```
YiseBoge, line 1: is this the best way to import the fonts ?

I have seen some people define them i think in the tailwind configs. might be worth looking into

Web/starter_project_vanguard/tailwind.config.js
```diff
@@ -9,17 +9,22 @@ module.exports = {
   theme: {
     extend: {
       colors: {
+        button_color: '#0263FF',
```
YiseBoge, line 12: should the button color be different from the primary color?
you can look at the design and if that is the color we are using primarily, i say turn the primary color to that

Web/starter_project_vanguard/tailwind.config.js
```diff
@@ -9,17 +9,22 @@ module.exports = {
...
         primary: '#264FAD',
         accent: '#FF9F43',
         danger: '#b91c1c',
         'primary-text': '#353535',
         'secondary-text': '#717171',
+
```
YiseBoge, line 18: [very-nit] no need for this change

YiseBoge (changes requested):
I have left some comments but the biggest issue is of not ignoring .next. let's figure that out and let me know if you need help too

Adona12:
The start of the  picture and the text should align. The text is centered vertically but that's not the case on the design. You should also check the font size of "Africa to silicon valley". It should be the same as the figma design.

PR #29, opened 2023-04-23
feat(web-barbies): Add team card component

hello-world-was-taken:
image


Web/starter_project_barbies/components/team/TeamCard.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react'
+import { BsFacebook, BsLinkedin, BsInstagram } from 'react-icons/bs'
+import Image from "next/image";
+
+type TeamCardProps = {
```
YiseBoge, line 5: i would recommend defining a member object and taking that as a prop than taking each one by one

so then you can have a list of member objects in the container and pass them into this component in a loop

Web/starter_project_barbies/components/team/TeamCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+  linkedin,
+  instagram,
+}: TeamCardProps) => {
+  return (
+    <div className="team-card box-content font-sans w-72 mt-4 text-center shadow-md shadow-gray-200 rounded-md py-2 m-auto">
+      <div className="team-card__image container overflow-hidden my-4">
+        <Image className="rounded-full mx-auto w-32" width={200} height={200} src={image} alt={name} />
+      </div>
+
+      <div className="team-card__content m-4">
```
YiseBoge, line 30: i don't think you need to prefix everything with team-card
you can just define the team card scss file and only import it in this component

YiseBoge: actually idk where these classes are being used 
are they only for semantics?

Web/starter_project_barbies/components/team/TeamCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+type TeamCardProps = {
+  name: string
+  jobTitle: string
+  description: string
+  image: string
+  facebook: string
+  linkedin: string
+  instagram: string
+}
```
Adona12, line 13: You should put the types you create in the "types" folder.

Web/starter_project_barbies/components/team/TeamCard.tsx
YiseBoge: and i would also rename the component as team member card instead of team card

Web/starter_project_barbies/components/team/TeamCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+        </p>
+        <p className="team-card__description text-sm mt-5 mb-3 text-gray-400 font-light px-4 tracking-wide">
+          {description}
+        </p>
+      </div>
+
+      <hr className="h-0.5 border-t-0 bg-neutral-100 opacity-100 dark:opacity-50 block m-auto w-5/6" />
+
+      <div className="team-card__social flex-row mt-3">
+        <div className="team-card__social__icon inline-block w-1/3">
```
YiseBoge, line 45: can you do these in a loop?
i think the social media fields list can be a part of the team member object defined to be a list of links (each link having an icon and a url)

implement it assuming not everyone has the same set of links (even if it is realistically not the case most of the time)

YiseBoge (approved):
Overall I like this but I have left some small comments to fix before merging
if it is working as it is, i like that you did not need to custom CSS to implement the look you wanted
Good job

Web/starter_project_barbies/components/team/TeamCard.tsx
hello-world-was-taken: No, they are not getting used now. It was in case we need to add custom css later on.

Web/starter_project_barbies/components/team/TeamCard.tsx
hello-world-was-taken: > 

I did that on the team members page. Should I change it?

hello-world-was-taken: > i would recommend defining a member object and taking that as a prop than taking each one by one
> 
> so then you can have a list of member objects in the container and pass them into this component in a loop

I did that on the team members page. Should I change it?

Web/starter_project_barbies/components/team/TeamCard.tsx
hello-world-was-taken: @YiseBoge I have tried to make it as you suggested, but since we are using React-Icons I wasn't able to just pass icon string (like in the ones font-awesome provides.) If you have another way of going about it, let me know.

Web/starter_project_barbies/components/team/TeamCard.tsx
YiseBoge: you can rename it to TeamMemberCard
and the team members page i assume would be team-members

Web/starter_project_barbies/components/team/TeamCard.tsx
YiseBoge: well if you are not using it i say get rid of it

Web/starter_project_barbies/components/team/TeamCard.tsx
YiseBoge: ok you don't need to pass strings then, pass the element as an object

YiseBoge: alternatively you can also just keep your own key value pair of which icon is for which social media link (assuming we want the social media to come from database and is savable)

Web/starter_project_barbies/types/team/TeamMemberCardProps.ts
```diff
@@ -0,0 +1,9 @@
+import { SocialMediaLink } from './SocialMediaLink'
+
+export type TeamMemberCardProps = {
```
YiseBoge, line 3: i know @Adona12 told you to do this but i prefer keeping prop definitions right above the component definition

YiseBoge: I would also still like you to take in just one TeamMember object that is defined here instead 
so basically I think this definition is more for a TeamMember object and the page props definition will take it as an object (which would correspond better with what you would find in the database)

Web/starter_project_barbies/types/team/SocialMediaLink.ts
YiseBoge, line 1: I think it would be fine to have teams.ts file in the types and have multiple type definitions in there
no need for a folder

Web/starter_project_barbies/components/team/TeamMemberCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <div className="team-card__social flex-row mt-3">
+        {links && links.map((link, index) => (
+        <div key={index} className="team-card__social__icon inline-block w-1/3">
+          <a
+            className="inline-block"
+            href={link.url}
+            target="_blank"
+            rel="noreferrer"
+          >
+            {link.type === 'facebook' && <BsFacebook className="fill-gray-400 w-6 h-6" />}
```
YiseBoge, line 42: yeah i don't think this is nice 
we can have a key value pair that holds it like `{facebook: <BsFacebook className="fill-gray-400 w-6 h-6" />, linkedin:...}` and index that using the link.type
and for any link type that is not included, you can have a default one in there too, and refer to it with something like this
`linkIcons[link.type] || inkIcons['default']` which just uses the default one if the first one is undefined

YiseBoge (approved):
I have left a few more comments
and let me remind to resolve comments you know you have fixed

Web/starter_project_barbies/types/team/TeamMemberCardProps.ts
hello-world-was-taken: > i know @Adona12 told you to do this but i prefer keeping prop definitions right above the component definition

So, should I change it back?

Web/starter_project_barbies/components/team/TeamMemberCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const SocialMediaIcon: {[index: string]:any} = {
+  "facebook": <BsFacebook className="fill-gray-400 w-6 h-6" />,
+  "linkedin": <BsLinkedin className="fill-gray-400 w-6 h-6" />,
+  "instagram": <BsInstagram className="fill-gray-400 w-6 h-6" />,
+  "default": <BsEmojiSmileUpsideDown className="fill-gray-400 w-6 h-6" />
+}
+
+const TeamMemberCard: React.FC<TeamMemberCardProps> = ({
+  teamMember
```
YiseBoge, line 18: you can destruct the teamMember here to avoid having to use teamMember. multiple times

Web/starter_project_barbies/components/team/TeamMemberCard.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+import { BsFacebook, BsLinkedin, BsInstagram, BsEmojiSmileUpsideDown } from 'react-icons/bs';
+import Image from "next/image";
+import { TeamMember } from "@/types/team";
+
+export type TeamMemberCardProps = {
+  teamMember: TeamMember
+}
+
+const SocialMediaIcon: {[index: string]:any} = {
```
YiseBoge, line 10: let's use cammelCase and avoid any

Web/starter_project_barbies/components/team/TeamMemberCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+  "facebook": <BsFacebook className="fill-gray-400 w-6 h-6" />,
+  "linkedin": <BsLinkedin className="fill-gray-400 w-6 h-6" />,
+  "instagram": <BsInstagram className="fill-gray-400 w-6 h-6" />,
+  "default": <BsEmojiSmileUpsideDown className="fill-gray-400 w-6 h-6" />
+}
+
+const TeamMemberCard: React.FC<TeamMemberCardProps> = ({
+  teamMember: {name, jobTitle, description, profileImg, socialMediaLinks }
```
YiseBoge, line 18: also this

Web/starter_project_barbies/types/team.ts
ruthgdb, line 1: Move this file to a /team folder

Web/starter_project_barbies/components/team/TeamMemberCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+          alt={name}
+        />
+      </div>
+
+      <div className="m-4">
+        <h3 className="text-2xl m-3 font-extrabold tracking-widest uppercase">
+          {name}
+        </h3>
+        <p className="m-3 tracking-widest text-base uppercase">{jobTitle}</p>
+        <p className="text-sm mt-5 mb-3 text-gray-400 font-light px-4 tracking-wide">
```
ruthgdb, line 42: [nit] its best if you use the colors you've already defined in your tailwind.config file to make it more uniform.

Web/starter_project_barbies/components/team/TeamMemberCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {
+  BsFacebook,
+  BsLinkedin,
+  BsInstagram,
+  BsEmojiSmileUpsideDown,
+} from 'react-icons/bs'
+import Image from 'next/image'
+import { TeamMember } from '@/types/team'
+
+export type TeamMemberCardProps = {
```
ruthgdb, line 11: If you've already defined its type and if your not planning on adding another property to this, you can just use the TeamMember type you imported. Theres no need to define another type here.

ruthgdb (changes requested):

sam5235 (approved):

PR #31, opened 2023-04-23
feat(Mobile-DartKnights): Implement Intro Pages for Dartknight Mobile 

Yared-betsega:
Team: Mobile

Team Name: DartKnights

Task name: Implement Splash screen, Onboarding 1, Onboarding 2

Notice: I didn't create a separate screens for Onboarding 1 and Onboarding 2, Instead I created a single widget and used state management to change the color of the icons. I did this because the two pages are completely the same and the only thing that is different is the three dots. Here is the only difference between them.
image

Screenshots:
1. Splash Screen
image

2. Onboarding_1 screen
image

3. Onboarding_2 screen
image


YiseBoge:
just looking at the PR description, I would assume that this PR addresses too many things (if it is the case that each one is a small page, then come up with a compressive message like `implement the intro pages for...` to incorporate everything and see if that is a big or small scope - there should be no `and`s in a commit message)

and looking at the changes count, i am now convinced this is doing a lot of things at once (that is not a good practice especially when you don't have really senior people to approve it)

Redi22:
There are changes committed to the other group's folder. They need to be untracked.

Redi22:
@YiseBoge   The Flutter cookbook suggests using the assets I think. I personally prefer the assets as well.  The other comments are good and need corrections though.

Yared-betsega:
> @YiseBoge The Flutter cookbook suggests using the assets I think. I personally prefer the assets as well. The other comments are good and need corrections though.

I have already changed the code based on @YiseBoge comment. Now, shall I change it back to assets?

Redi22:
> > @YiseBoge The Flutter cookbook suggests using the assets I think. I personally prefer the assets as well. The other comments are good and need corrections though.
> 
> I have already changed the code based on @YiseBoge comment. Now, shall I change it back to assets?

It is fine, keep it. we will discuss this and give you guys a uniform way of using it.

Yared-betsega:
> > > @YiseBoge The Flutter cookbook suggests using the assets I think. I personally prefer the assets as well. The other comments are good and need corrections though.
> > 
> > 
> > I have already changed the code based on @YiseBoge comment. Now, shall I change it back to assets?
> 
> It is fine, keep it. we will discuss this and give you guys a uniform way of using it. However, I still see changes in other teams' folders. Could you please fix those?

I'm working on that.

Redi22:
Can you also make the commit messages descriptive? instead of "removed unnecessary commits" you could say "removed these folders" or "removed these files from the PR"  or sth like that.

PR #36, opened 2023-04-23
feat(web-barbies) : Add select blog tag component

TsehaynehGetaneh:
Add select tag component for blog post
  
- Figma Design : 
select-tag-figma

- implementation

select-tag-imp


Adona12 (approved):
I think it looks great

YiseBoge (commented):
this seems good but let's make sure we customize the UI of the component (it does the job now but it is not look like the design)
but that can be a job on another PR, let's just make sure we track that

PR #37, opened 2023-04-23
feat(web-barbies): Add team page

hello-world-was-taken:
image
^ DESKTOP VIEW

image
^ TAB SCREEN-ISH VIEW

image
^ MOBILE VIEW


Web/starter_project_barbies/pages/team/Team.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
+import TeamsImage from '../../public/teams_page_img.svg'
+import Image from 'next/image'
+import TeamMemberCard from "@/components/team/TeamMemberCard";
```
YiseBoge, line 4: this does not exist yet
so you have to make sure this does not merge before the other one does

Web/starter_project_barbies/pages/team/Team.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { TeamMemberCardProps } from "@/types/team/TeamMemberCardProps";
+
+const DummyDataGenerator = (): TeamMemberCardProps[] => {
+  const data:TeamMemberCardProps[] = []
```
YiseBoge, line 8: see I don't imagine it would be a good idea to have a list of props
I think if you have a TeamMember object defined separately (like mentioned in the other PR #29) you can have your data be a list of that

Web/starter_project_barbies/pages/team/Team.tsx
```diff
@@ -0,0 +1,67 @@
...
+  for (let i = 0; i < 10; i++) {
```
YiseBoge, line 9: why have this be a function?
why not define a dummy data json file and import that as a team member list? (you can put it inside a data folder in the root)

Web/starter_project_barbies/pages/team/Team.tsx
```diff
@@ -0,0 +1,67 @@
...
+        </div>
+        <div className="w-0 md:w-3/5 my-8">
+          <Image src={TeamsImage} alt="placeholder" />
+        </div>
+      </div>
+
+      <hr className="h-0.5 border-t-0 bg-neutral-100 opacity-100 dark:opacity-100 block w-5/6 my-8 m-auto" />
+
+      <div className="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4">
+        {DummyDataGenerator().map((data, index) => (
```
YiseBoge, line 59: [nit] i would use a proper name for the team members list here (when you end up using proper team members at the top, no need to rename this after)

Web/starter_project_barbies/pages/team/Team.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <TeamMemberCard key={index} name={data.name} jobTitle={data.jobTitle} description={data.description} image={data.image} links={data.links} />
```
YiseBoge, line 60: here if it took a team member object, it could easily be passed to it as `member={member}`

Web/starter_project_barbies/public/teams_page_img.svg
YiseBoge, line 1: I don't think this is the right place to have this photo
I would put it in a place like (`img/[feature-name]` - could be `/img/team` for this)

YiseBoge: and try to have a name for the file that is better descriptive (you would not need to call it teams or page if it is inside the subfolders, so describe what it's role is in that page)

and use kebab-case please

Web/starter_project_barbies/pages/team/Team.tsx
YiseBoge: The proffessional way to handle this would be to create a PR that defines the common types and components that both PR's use and merge than first so that the other PR's can work independently as much as possible

YiseBoge (changes requested):
i have left some comments about the structure (could relate to the TeamMemberCard PR too) but overall this seems nice to me

Web/starter_project_barbies/pages/team/Team.tsx
hello-world-was-taken: > this does not exist yet so you have to make sure this does not merge before the other one does

Since I changed the base branch for this PR, it merges to the one containing the component.

hello-world-was-taken: > The proffessional way to handle this would be to create a PR that defines the common types and components that both PR's use and merge than first so that the other PR's can work independently as much as possible

The other branch was created to create the common component and this branch merges to the one containing the common component, if it made sense.

Web/starter_project_barbies/data/team/data.json
YiseBoge, line 1: for consistency, i would recommend using a team.json file instead of a folder

Web/starter_project_barbies/public/image/team/teams_page_img.svg
YiseBoge, line 1: let's use kebab-case

YiseBoge (changes requested):
I have left some nits that would be nice to fix

PR #38, opened 2023-04-23
Kaleab.about-page

kkstarkk:
In this branch I have implemented the About Us Page. I have divided the page into 4 sections: Hero Section, Problems Section, 
Social Projects Section and Sessions Section.
**Hero Section:**
Figma
image
Implementation:
image

**Problems Section:**
Figma:
image
Implementation:
image

**Social Projects Section:**
Figma:
image

Implementation:
image

**Sessions Section:**
Figma:
image

Implementation:
image


Web/starter_project_barbies/components/about/HeroSection.tsx
```diff
@@ -0,0 +1,63 @@
+import Image from 'next/image'
+
+const HeroSection = ({}) => {
```
Adona12, line 3: Can you use syntax like React.FC to define a component? It is better to make it clear that we are looking at a component

Web/starter_project_barbies/public/about/africa.svg
Adona12, line 1: Let's have a clear folder structure. Images should be inside the img folder then we can have a subfolder inside it

Web/starter_project_barbies/components/about/HeroSection.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <div className="flex justify-center gap-10 flex-wrap">
+      <div className="flex flex-col justify-center">
+        <div className="font-bold md:text-2xl sm:text-xl lg:text-3xl xl:text-4xl p-3">
+          <span className="text-blue-500">Africa </span>to Silicon Valley
```
Adona12, line 8: Let's define colors globally(tailwind.config.js) so others can also use it.

Web/starter_project_barbies/public/about/amicopuzzle.png
Adona12, line 1: Use kebab case for everything except the components. You should separate the words with a dash

Web/starter_project_barbies/components/about/ProjectCard.tsx
```diff
@@ -0,0 +1,49 @@
+import { ProjectCardProps } from '@/types/AboutTypes'
+import Image from 'next/image'
+
+const ProjectCard = ({ project, isEven }: ProjectCardProps) => {
```
Adona12, line 4: Let's define props here above the component so it's more clear. I think it would be best to do it for all the components whose props are define on types folder

Web/starter_project_barbies/public/about/approach.svg
Adona12, line 1: I would suggest having subgroups for the images, you can put session related images in a session folder, same thing for projects and you can be creative and create folder for the others. The main thing is to divide them rather than having all this image in just one folder.

Web/starter_project_barbies/data/about.ts
YiseBoge, line 1: can we have this data be from a json file?
i woule personally prefer it as it lets us treat it similar to what we would find from backend

YiseBoge: it is fine to have multiple json files if you need

Web/starter_project_barbies/data/about.ts
```diff
@@ -0,0 +1,75 @@
+import { ProblemItem, Project, Session } from '@/types/AboutTypes'
+
+export const Projects: Project[] = [
+  {
+    image: '/about/hakimhub.png',
```
YiseBoge, line 5: none of the links should have the `/about` in them
you should decide that when you are rendering the page because it is common for all)
it also gives the flexibility to change that folder without having to modify all these links

Web/starter_project_barbies/data/about.ts
```diff
@@ -0,0 +1,75 @@
...
+    session: 'Technical Training',
+    description:
+      '6 days a week, 3 hours of lectures and practice sessions to improve technical problem-solving skills.',
+    image: '/about//technicaltraining.svg',
+  },
+  {
+    session: 'Q&As',
+    description:
+      'In Q&As, we get to know engineers, founders, and entrepreneurs from top tech companies. We see that they are normal people like us and we learn the best practices.',
+    image: '/about//qanda.svg',
```
YiseBoge, line 34: [nit] there is no need for `//` (it appears multiple places)

Web/starter_project_barbies/data/about.ts
```diff
@@ -0,0 +1,75 @@
...
+      'Africa’s future depends on innovation. Transformative technologies can drive rapid economic growth and lift millions of people out of poverty. However, university computer science education is misaligned with global market needs and fails to incorporate practice-based learning, leaving students unable to apply their skills in real-world contexts.',
+  },
+  {
+    icon: '/about/code.svg',
+    description:
+      'With few global tech companies on the continent, aspiring engineers don’t have access to experienced mentors, or the opportunity to work on products that operate at scale. Smart and ambitious students who could create life-changing technologies aren’t able to reach their potential.',
+  },
+]
+
+export const SolutionItemList: ProblemItem[] = [
```
YiseBoge, line 69: if the ProblemItem type is the same one we use for SolutionItem too it might be better to rename it to be inclusive

YiseBoge: alternatively, you can define a SolutionItem as 
`type SolutionItem = ProblemItem`
in your type definitions to have a copy of it that acts the same way but it has a clearer name

Web/starter_project_barbies/components/about/SessionList.tsx
YiseBoge, line 1: I think it might be fine to define the SessionCard component inside this file 
and I would call this component `SessionsList` (Sessions plural)
this is because SessionCard is only used in this component and it is better to define it here and not have to export it
They are also both small components so it won't be too much for one file

Web/starter_project_barbies/components/about/SessionCard.tsx
```diff
@@ -0,0 +1,20 @@
+import { SessionProps } from '@/types/AboutTypes'
```
YiseBoge, line 1: the prop type definition should follow a pattern of `[ComponentName]Props` so this should be `SessionCardProps`

Web/starter_project_barbies/components/about/ProblemItemCard.tsx
YiseBoge, line 1: this can be defined inside the ProblemSection file as it is only used in there and it is better not to have to export it

Web/starter_project_barbies/components/about/ProblemSection.tsx
YiseBoge, line 1: [nit] I would rename this to `ProblemsSection` because it holds multiple problems

Web/starter_project_barbies/components/about/ProblemItemCard.tsx
YiseBoge: and it is small enough to not be a problem too

Web/starter_project_barbies/components/about/ProblemSection.tsx
```diff
@@ -0,0 +1,46 @@
+import { ProblemItemList, SolutionItemList } from '@/data/about'
```
YiseBoge, line 1: variable names should follow `camelCase` format (these are not class definitions)

Web/starter_project_barbies/components/about/HeroSection.tsx
YiseBoge: i think you should be using the primary color here as well (make sure the primary is defined appropriately too)
because the current color feels a bit off in the UI

Web/starter_project_barbies/components/about/HeroSection.tsx
```diff
@@ -0,0 +1,63 @@
...
+              src="/about/education_process.png"
+              alt="education"
+              width={300}
+              height={50}
+            />
+            <div className="text-white text-xs sm:text-sm md:text-md lg:text-lg xl:text-xl text-center absolute inset-0 mx-auto my-auto w-36 h-fit">
+              The Education Process
+            </div>
+          </div>
+          <div className="relative  flex-1">
```
YiseBoge, line 40: [very-nit] `relative  flex-1` has an extra space in between

Web/starter_project_barbies/components/about/HeroSection.tsx
```diff
@@ -0,0 +1,63 @@
...
+          students to create digital solutions to Africa’s most pressing
+          problems.
+        </p>
+      </div>
+      <div className="gap-2">
+        <div className="mb-10 font-bold p-3 text-xs sm:text-sm md:text-md lg:text-lg xl:text-xl">
+          Group Activities
+        </div>
+        <div className="flex gap-2 flex-wrap">
+          <div className="relative flex-1 ">
```
YiseBoge, line 29: [very-nit] `relative flex-1 ` has an extra space in the end

Web/starter_project_barbies/components/about/ProjectCard.tsx
YiseBoge, line 4: instead of `isEven`, i would rather take in `shouldReverseFlex` or sth
basically something that describes the UI because the only reason we need the evenness is to check if it should reverse (we should not care about why it should reverse in here)

YiseBoge: in this case of course all you'd need to do is rename the variable
but i just wanted to explain the concept

Web/starter_project_barbies/components/about/ProjectCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+          }`}
+        >
+          <Image
+            className="flex-initial"
+            src="/about/github.svg"
+            width={20}
+            height={20}
+            alt="github"
+          />
+          <Image
```
YiseBoge, line 36: a lot of these images can be replaced with icons 
better to do that if possible

Web/starter_project_barbies/components/about/ProjectCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+      className={`flex flex-wrap justify-around ${
+        !isEven ? 'flex-row-reverse' : ''
+      }`}
+    >
+      <Image
+        className="flex-initial"
+        src={project.image}
+        width={700}
+        height={200}
+        alt="hakimhub"
```
YiseBoge, line 16: [very-nit] i think you don't need to condense the alt too much, you can just say Hakim Hub or sth

Web/starter_project_barbies/components/about/ProjectCard.tsx
```diff
@@ -0,0 +1,49 @@
+import { ProjectCardProps } from '@/types/AboutTypes'
+import Image from 'next/image'
+
+const ProjectCard = ({ project, isEven }: ProjectCardProps) => {
+  return (
+    <div
+      className={`flex flex-wrap justify-around ${
```
YiseBoge, line 7: [nit] can you try this format:

```
className={{
  flex,
  flexWrap,
  justifyAround,
  flexRowReverse: isEven
}}
// not entirely sure if the keys should be in camelCase or kebab-case with quotes
```

it is an alternative approach that might be cleaner

Web/starter_project_barbies/components/about/ProjectCard.tsx
YiseBoge, line 1: The project card could be defined in the same file as the project list as that is the only place it is used
it might be better to not have to export it

Web/starter_project_barbies/components/about/ProjectList.tsx
```diff
@@ -0,0 +1,19 @@
+import { Projects } from '@/data/about'
```
YiseBoge, line 1: variables should be in camelCase

Web/starter_project_barbies/components/about/ProjectList.tsx
YiseBoge, line 1: i would rename this to `ProjectsList` (it holds multiple projects)

Web/starter_project_barbies/public/about/hakimhub.png
YiseBoge, line 1: [nit] could be named `hakim-hub`

YiseBoge (changes requested):
The implementation seems nice overall
i have left some comments (most of them are minor)
some are related to having too many files (be a bit more careful with their naming standards too)

sam5235 (approved):

YiseBoge:
let's please rename the PR title to follow the conventional commit rules
[nit] `kaleab.about-page` is a better name for the branch for next time

Web/starter_project_barbies/components/about/ProjectsList.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <IconContext.Provider value={{ className: 'text-primary' }}>
+            <IoMdOpen size="1.25em" />
+          </IconContext.Provider>
+        </div>
+      </div>
+    </div>
+  )
+}
+
+const ProjectsList = () => {
```
YiseBoge, line 53: [nit] please use React.FC syntax to define components

Web/starter_project_barbies/data/about.tsx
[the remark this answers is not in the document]
YiseBoge: it might be fine to do this in another PR if you want to, but it will be necessary for the next task

YiseBoge (changes requested):
i have left some nits that i think should take you short time to finish
and please lint the project before pushing and merge any conflicts if they arise

Web/starter_project_barbies/components/about/ProblemsSection.tsx
```diff
@@ -0,0 +1,61 @@
+import problemItemList from '../../data/about/problem-items.json'
+import solutionItemList from '../../data/about/solution-items.json'
+
+import { ProblemItem } from '@/types/about'
+import Image from 'next/image'
+
+interface ProblemItemCardProps {
+  Item: ProblemItem
```
YiseBoge, line 8: [nit] I would rename `Item` to `problemItem` as that is what it seems to be and you should stay consistent with the camelCase

Web/starter_project_barbies/components/about/ProblemsSection.tsx
```diff
@@ -0,0 +1,61 @@
...
+          src="/img/about/problems/amico-puzzle.png"
+          alt=""
+          width={500}
+          height={100}
+        />
+        <div className="flex-initial order-2">
+          <div className="font-bold text-center md:text-left lg:text-5xl md:text-4xl sm:text-3xl xs:text-2xl max-w-lg">
+            How We Are <span className="text-primary">Solving it</span>
+          </div>
+          {solutionItemList.map((Item, index) => (
```
YiseBoge, line 52: again, lets stick to camelCase for the variable name, and let's make sure we rename it appropriately too
Item here is being used to represent both solutionItem and problemItem above in the same file. This can be confusing

Web/starter_project_barbies/components/about/ProjectsList.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+interface ProjectCardProps {
+  project: Project
+  reverseFlex: boolean
+}
+
+const ProjectCard: React.FC<ProjectCardProps> = ({ project:{image, name, description}, reverseFlex }) => {
+  return (
+    <div
+      className={`flex flex-wrap justify-around ${
```
YiseBoge, line 16: [very-nit] one alternative way to do this if you want is
```
className={{
  flex,
  flexWrap,
  justifyAround,
  flexRowReverse: reverseFlex
}}
```
(not entirely sure if the class names need to be camelCase or 'kebab-case' as strings)

No need to change now, just wanted to show you an alternative way if you want. this way could be useful when you have class names that will be needed only upon a condition (the ones that should always be there will just be listed as if their condition is true). and it would be cleaner expecially if there are multiple of them

Web/starter_project_barbies/pages/about/index.tsx
```diff
@@ -0,0 +1,20 @@
+import HeroSection from '../../components/about/HeroSection'
```
YiseBoge, line 1: [nit] i think you can use `@/components/...` for your paths (this might apply elsewhere too)

YiseBoge (changes requested):
This looks really nice
i have left some small nits on it tho, should be 5 minute fix

Web/starter_project_barbies/components/about/ProblemsSection.tsx
kkstarkk: i named it Item because the prop could be problemItem or solutionItem. shall i rename the component to ItemCard instead??

kkstarkk: i named it Item because the prop could be problemItem or solutionItem. shall i rename the component to ItemCard instead??

Web/starter_project_barbies/components/about/ProjectsList.tsx
```diff
@@ -0,0 +1,71 @@
...
+        className={`mx-5 ${!reverseFlex ? 'lg:text-right' : 'lg:text-left'}`}
+      >
+        <div className="text-primary">Social Project</div>
+        <div className="text-primary font-bold text-3xl mt-2 mb-5">
+          {name}
+        </div>
+        <div className="lg:max-w-sm text-gray-600">{description}</div>
+        <div
+          className={`flex gap-5 mt-5 ${
+            !reverseFlex ? 'lg:justify-end' : 'lg:justify-start'
```
YiseBoge, line 37: [nit] this is fine for now but generally it is better to not have the negative in the conditional, I'd rather you swap the justify start and justify end instead of adding a not on the reverseFlex
I think you've done that other places too

YiseBoge (approved):
This looks good
Well done

PR #40, opened 2023-04-23
feat(web-iconics) : implement footer component

Aklile-Yilma:
footer-desktop
footer-mobile


Web/starter_project_iconics/public/helping-a-partner-rafiki.jpg
YiseBoge, line 1: wait why are there two of these same image?
if you are not using one, I say get rid of it

YiseBoge: and i don't think this is the right folder for it too
i think it should atleast be inside the `img/[your feature]` folder to not have a mess with other images (for this the feature could be `layout` or sth including the navbar idk)

Web/starter_project_iconics/components/layout/Footer.tsx
```diff
@@ -0,0 +1,116 @@
...
+  AiOutlineInstagram,
+  AiOutlineTwitter,
+} from 'react-icons/ai'
+import { MdCopyright, MdOutlineFacebook } from 'react-icons/md'
+
+const Footer: React.FC = () => {
+  const navigation = [
+    {
+      header: 'Links',
+      links: [
```
YiseBoge, line 14: I assume this navigation list is also useful in the navbar. it would be nice if we could share the same list for the two
i could recommend maybe exporting from one and importing in the other

Web/starter_project_iconics/components/layout/Footer.tsx
```diff
@@ -0,0 +1,116 @@
...
+                    )}
+                  >
+                    <a className="font-light" href={link_item.name}>
+                      {link_item.name}
+                    </a>
+                  </li>
+                ))}
+              </ul>
+            </div>
+          ))}
```
YiseBoge, line 96: i like this 👍

Web/starter_project_iconics/components/layout/Footer.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <p>
+          <MdCopyright className="inline mr-2" />
+          2020 Africa to Silicon Valley, Inc. All right reserved.
+        </p>
+        <div className="flex gap-4 text-2xl md:text-4xl ">
+          <AiOutlineTwitter />
+          <MdOutlineFacebook />
+          <AiFillYoutube />
+          <AiFillLinkedin />
+          <AiOutlineInstagram />
```
YiseBoge, line 109: would be nice if we could use a loop for this too (you could have a list of social links that has an icon and a url as well)

YiseBoge (changes requested):
Overall this would have been a really good PR but you kinda made a mess of the other team's projects
I think you should fix all that first (untrack other team changes), then the rest is minor comments

Web/starter_project_iconics/types/nav-item.ts
YiseBoge, line 1: [nit] might be better to stay consistent with the file naming of `layout` (to stay the same as the folder in the compoenents, and i would say the same for the img folder too btw)

Adona12 (dismissed):
LGTM

ruthgdb (approved):
LGTM

PR #42, opened 2023-04-23
feat(web-iconics): add configuration for font-styles

Aklile-Yilma:
I also i want to add configuration for font-sizes, do you think it's necessary to add them.

### DIFF
--- a/Web/starter_project_iconics/styles/globals.css
+++ b/Web/starter_project_iconics/styles/globals.css
@@ -1,6 +1,13 @@
+@import url('https://fonts.googleapis.com/css2?family=Montserrat:wght@200;300;400;600;700;800&family=Poppins:ital,wght@0,200;0,300;0,400;0,700;0,800;1,600&display=swap');
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+@layer base {
+  html {
+    font-family: 'Poppins', sans-serif;
+  }
+}
 
 :root {
   --foreground-rgb: 0, 0, 0;
--- a/Web/starter_project_iconics/styles/globals.css
+++ b/Web/starter_project_iconics/styles/globals.css
@@ -1,13 +1,6 @@
-@import url('https://fonts.googleapis.com/css2?family=Montserrat:wght@200;300;400;600;700;800&family=Poppins:ital,wght@0,200;0,300;0,400;0,700;0,800;1,600&display=swap');
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
-
-@layer base {
-  html {
-    font-family: 'Poppins', sans-serif;
-  }
-}
 
 :root {
   --foreground-rgb: 0, 0, 0;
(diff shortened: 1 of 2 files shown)


Web/starter_project_iconics/styles/globals.css
```diff
@@ -1,7 +1,14 @@
...
 @tailwind components;
 @tailwind utilities;
 
+@layer base {
+  html {
+    font-family: 'Poppins', sans-serif;
+  }
+}
+
 :root {
```
Adona12, line 12: I would suggest removing css configurations that you're not using. I belive these are generated by default. If you're going to use them, it's okay you don't have to remove it but if not I don't see the need to have it.

Web/starter_project_iconics/styles/globals.css
```diff
@@ -1,7 +1,14 @@
+@import url('https://fonts.googleapis.com/css2?family=Montserrat:wght@200;300;400;600;700;800&family=Poppins:ital,wght@0,200;0,300;0,400;0,700;0,800;1,600&display=swap');
```
YiseBoge, line 1: i am not sure if this is the right place to import this
i think there might be a javascript way of doing it without having to import the whole font in the glbal css
please check if that is the case but otherwise it could be fine

Web/starter_project_iconics/styles/globals.css
```diff
@@ -1,7 +1,14 @@
...
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
 
+@layer base {
+  html {
+    font-family: 'Poppins', sans-serif;
```
YiseBoge, line 8: is this the case for everyone ?
you can check the designs to make sure this applies for every page

YiseBoge: personally, i would prefer if there is a way to only do configurations in tailwind and it's own text styles - same way we can override the primary colors and so on (not sure if that is an option i think it is worth checking)

YiseBoge (changes requested):
I am a bit unsure of doing changes in the global css, so i would recommend doing some checks to only customize tailwind. 
if that does not end up as the best way, then it should be fine tho

Adona12:
> I also i want to add configuration for font-sizes, do you think it's necessary to add them.

Web/starter_project_iconics/tailwind.config.js
```diff
@@ -20,6 +21,12 @@ module.exports = {
         'gradient-conic':
           'conic-gradient(from 180deg at 50% 50%, var(--tw-gradient-stops))',
       },
```
Adona12, line 23: If this is the default config, I think it's better to remove it

Web/starter_project_iconics/tailwind.config.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
     extend: {
       colors: {
         primary: '#264FAD',
+        secondary: '#E1E7EC',
         accent: '#FF9F43',
         danger: '#b91c1c',
         'primary-text': '#353535',
```
Adona12, line 16: Why do we need the quotation on the name?

Adona12 (approved):
Looks good👍

YiseBoge (approved):
Looks good to me

PR #43, opened 2023-04-24
feat(web-barbies): Create blog detail screen

nawrazi:


Web/starter_project_barbies/styles/blog/blog-detail.module.css
```diff
@@ -0,0 +1,10 @@
+@import url('https://fonts.googleapis.com/css2?family=IM+Fell+French+Canon&family=Montserrat:wght@200&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Montserrat:wght@300&display=swap');
+
+.cannon {
+  font-family: 'IM FELL French Canon';
+}
+
+.monteserrat {
+  font-family: 'Montserrat';
+}
```
Adona12, line 10: There's another way to set the font-family. You can define it on the tailwind.config file instead of doing it on the css file.

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
```diff
@@ -0,0 +1,137 @@
+import BlogCard from '@/components/blog/BlogCard'
+import fonts from '@/styles/blog/blog-detail.module.css'
+import { Blog } from '@/types/blog'
+import Image from 'next/image'
+
+interface BlogDetailProps {
+  blog: Blog
+}
+
+interface BlogAuthorDetailProps {
```
YiseBoge, line 10: [nit] this is fine but i would prefer if each prop is above it's own component for a bit more consistency and clarity

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
```diff
@@ -0,0 +1,137 @@
...
+  author: {
+    name: string
+    email: string
+    image: string
+    profession: string
+    userName: string
+  }
```
YiseBoge, line 17: for this definition, I think the Blog interface has its own author definition right?
you can use `author: Blog['author']` to refer to the author definition inside of the blog definition
this makes it so that there is only one place to modify the author definition (and avoid duplicate wirk)

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
```diff
@@ -0,0 +1,137 @@
...
+}
+
+interface BlogContentProps {
+  content: string
```
YiseBoge, line 21: [nit] you can use something similar as above for this too
allowing us to change the type of the content if we want to (maybe we want to make it an HTML entity instead of a string or sth)

but to be honest i don' think we will change it

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
```diff
@@ -0,0 +1,137 @@
...
+}
+
+interface RelatedBlogsProps {
```
YiseBoge, line 24: [nit] for this it is fine to reuse the BlogDetailProps as a related blog is really also a blog too

Web/starter_project_barbies/styles/blog/blog-detail.module.css
YiseBoge, line 1: can we use scss instead of CSS
i understand that might not have been communicated properly but let's try to do that
i will keep telling others to stay consistent too

YiseBoge: I like the file structure tho 👍

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
```diff
@@ -0,0 +1,137 @@
...
+      {paragraphs}
+    </div>
+  )
+}
+
+const RelatedBlogs: React.FC<RelatedBlogsProps> = ({ blog }) => {
+  // Fetch related blogs
+  let relatedBlogs: Blog[] = new Array(3).fill(blog);
+
+  let blogs = []
```
YiseBoge, line 117: [nit] i would rename this to 'relatedBlogComponents' or sth 
and not lose the idea that these are still related blogs

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
```diff
@@ -0,0 +1,137 @@
...
+      {/* Related blogs */}
+      <div className='flex items-center justify-center pb-20'>
+        <RelatedBlogs blog={blog} />
+      </div>
+
+    </div>
+  )
+}
+
+const BlogAuthorDetail: React.FC<BlogAuthorDetailProps> = ({ author }) => {
```
YiseBoge, line 70: [nit] if you want to, you can receive the author in a broken down way by using something like `author: {name, image, ...}` in place of the author part 
this lets you avoid using `author.` everywhere

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
```diff
@@ -0,0 +1,137 @@
...
+      {line}
+    </p>
+    paragraphs.push(paragraph)
+  }
+
+  return (
+    <div className='mx-96'>
+      {paragraphs}
+    </div>
+  )
```
YiseBoge, line 110: [nit] this seems good now, if we end up using a text editor to add the blog, the content may not be saved in the same way so let's make sure to update it when that is done (perhaps talk with whoever is doing that too)

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
```diff
@@ -0,0 +1,137 @@
...
+        <RelatedBlogs blog={blog} />
+      </div>
+
+    </div>
+  )
+}
+
+const BlogAuthorDetail: React.FC<BlogAuthorDetailProps> = ({ author }) => {
+  return (
+    <div className={"text-xs font-extralight " + fonts.monteserrat}>
```
YiseBoge, line 72: [nit] i would use parameterized strings for such things (\`text-xs font-extralight ${fonts.monteserrat}\`)

YiseBoge (approved):
I generally like this a lot 
I have left quite a few nits to solve before merging tho
(also let's remember adona's comment)

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  return (
+    <div className='mx-96'>
+      {paragraphs}
+    </div>
+  )
+}
+
+const RelatedBlogs: React.FC<RelatedBlogsProps> = ({ blog }) => {
```
Adona12, line 113: I would move this component into a separate file. It's a component that could possibly be used somewhere else maybe not in this scope but it's just a suggestion, it's fine even if you don't change it

Web/starter_project_barbies/pages/blog/BlogDetail.tsx
nawrazi: @Adona12 has asked for the whole component to be moved to its own file. I think it's better to take the interface too instead of importing BlogDetailProps from here, even though you are right about the repetitiveness.

Web/starter_project_barbies/styles/blog/blog-detail.module.css
nawrazi: Yeah I guess you're right. Will change it.

Adona12 (approved):

PR #55, opened 2023-04-24
implement blog card

TseganeshYifruBerta:


TseganeshYifruBerta:
image

Web/starter_project_vanguard/components/blog/avatar.tsx
ruthgdb, line 1: General components like this should not be specified in the blogs folder since they can possibly used somewhere else. Components in the blog folder should only contain components specific to the blog.

Web/starter_project_vanguard/components/blog/name.tsx
ruthgdb, line 1: The whole component can be included in your blog card, specially if your not using it anywhere else.

ruthgdb (changes requested):

PR #67, opened 2023-04-25
Define custom button

jebessaD:
Here are the sample buttons
buttons photo

The code
buttons code


YiseBoge (approved):
these seem fine 
let us make sure it is well communicated with the team members to use them

PR #69, opened 2023-04-25
Update footer - change links and button class

jebessaD:
update footer

**I altered the following codes:**

1. separate the navigation and social links from the footer.txs
2. convert the button to the newly defined button class
3. Modify the Iteration Method on Links 
4. Rename the file "footer.tsx" to "Footer.tsx"
5. relocate the image 
6. add responsiveness


Web/starter_project_vanguard/components/layout/links/navigationLink.tsx
YiseBoge, line 1: ok I see what you are doing here, saving the data elsewhere and importing it
but in this case, this file is not a component (TSX files are for components and pages)

YiseBoge: Additionally, file names should be kebab-case except for components which should be PascalCase, never camelCase

YiseBoge: these files are holding `data` and `types` definitions, which belong in their own folders separately (not inside components)

YiseBoge: actually tho, i don't think any of these data need to stay inside data, i think it was fine like before to keep them within components (as that is the only place they are relevant and they are not dependent on other things or are presenting extra information, they are just UI items)
i would recommend sticking with that

YiseBoge: My previous comment was only about importing the already existing definition of the nav links variable from the header component in order to reuse it

Web/starter_project_vanguard/components/layout/Footer.tsx
```diff
@@ -0,0 +1,60 @@
+import Link from 'next/link'
+import empowered from 'public/img/footer/empowered.png'
+import Image from 'next/image'
+import { linkItems } from './links/navigationLink'
+import { socialMedialinks } from './links/socialMediaLinks'
+
+const Footer: React.FC = () => {
+  return (
+    <footer className="mt-auto border-y border-text-secondary sm:px-6 px-12">
+      <div className="px-6 sm:flex justify-between items-center space-y-12 sm:space-y-0 divide-y divide-y-reverse sm:py-12 py-3 sm:divide-y-0  sm:space-x-12">
```
YiseBoge, line 10: [very-nit] `sm:divide-y-0  sm:space-x-12` has two spaces in between

Web/starter_project_vanguard/components/layout/Footer.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </div>
+
+      <div className="md:border-t border-neutral-200 sm:flex mx-auto md:justify-between py-8">
+        <p className="text-sm text-secondary-text py-4 sm:py-0 ">
+          <span className="text-xl px-1">&copy;</span> 2023 Africa to Silicon
+          Valley,Inc. All right reserved.
+        </p>
+
+        <div className="text-secondary-text flex align-middle justify-center space-x-6 text-xl">
+          {socialMedialinks.map((socialMediaLink, index) => {
```
YiseBoge, line 47: [nit] you can destruct the `socialMediaLink` here like `...socialMediaLink: {url, icon}...` to avoid using `socialMediaLink.` multiple times

Web/starter_project_vanguard/components/layout/links/socialMediaLinks.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface iconLink {
+    icon: any
+    url : string
+  }
+  
+  export const socialMedialinks : iconLink[] = [
+
+    {
+        icon: <FaTwitter />,
```
YiseBoge, line 11: using components in a list is fine when having a list inside another component, but if it were defined inside the data folder, it would required a workaround

Web/starter_project_vanguard/components/layout/links/socialMediaLinks.tsx
```diff
@@ -0,0 +1,31 @@
+import { FaFacebook, FaInstagram, FaLinkedin,FaTwitter,FaYoutube} from 'react-icons/fa'
+
+interface iconLink {
```
YiseBoge, line 3: [nit] type names should be in PascalCase (but actually you can also leave these type definitions out if you don't want them - it is not a bad thing to have them tho)

YiseBoge (changes requested):
i like the implementation but I have left some comments that would be nice to fix before merging
it would also be nice if you run lint and prettier to your project to help with some inconsistencies

YiseBoge (changes requested):
i like the implementation but I have left some comments that would be nice to fix before merging
it would also be nice if you run lint and prettier to your project to help with some inconsistencies

YiseBoge (approved):
This looks good

PR #84, opened 2023-04-27
feat(web-barbies) : Create a Navigatiion bar

kidus101:
Add Navigation Bar

 - The Navigation bar contains  a logo , different tabs and Login and Donate button to the right.
 
 Figma Design :
navbar figma design

Implementation:
navbar implementation design


Web/starter_project_barbies/pages/index.tsx
Yared04, line 1: If you are not responsible for implementing the homepage you should not make changes to this file as it will create conflicts for whoever else is responsible.

kidus101: alright I get it . Thank You.

Web/starter_project_barbies/public/img/A2SV Logo.jpeg
ruthgdb, line 1: use kabab-case for all you images

ruthgdb (changes requested):
Looks good overall. I've left some comments for you to address.

PR #89, opened 2023-04-27
feat(web-iconics): add personal info page component in profile

blenassefa2:
In this branch the main component of the personal info page is created. 
I just used a form with three questions (name, email address and your photo) and made the UI resemble the figma
 
The figma looks as follows:
image

The implemented  page looks as follows:
image


Web/starter_project_iconics/pages/profile/my-info.tsx
ruthgdb, line 1: If you add a page folder (in this case 'profile') under pages folder, it should contain and index.tsx file since thats the file it looks to render when you go to /profile route.

Look at [this](https://nextjs.org/docs/routing/introduction) doc for more info.

Web/starter_project_iconics/components/profile/ProfileLayout.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const [activeLink, setActiveLink] = useState(currentPage)
+
+  const handleLinkClick = (linkText: string) => {
+    setActiveLink(linkText)
+  }
+
+  return (
+    <div className="min-h-screen bg-white flex p-10 flex-col">
+      <div className="flex   py-4">
+        <h1 className="text-4xl font-bold text-gray-900">Profile</h1>
```
ruthgdb, line 28: This should be one of your primary colors if its a header color. You should add it to the config file (if its not there already) and use that

Web/starter_project_iconics/components/profile/ProfileLayout.tsx
```diff
@@ -0,0 +1,83 @@
...
+      </div>
+      <div className="flex flex-col md:flex-row justify-center items-center bg-white pb-0">
```
ruthgdb, line 30: You can store this in array and loop over it too. The code seems a bit redundant.

PR #111, opened 2023-05-02
fix bugs

tamiratdereje:
implement post, edit article ui with fields of title, subtitle, content, and tags.
tammmmmmmmmmmmmm


Mobile/starter_project_dartsmiths/lib/core/utils/article_page_styles.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:flutter/material.dart';
+
+
+class BlogFontConstant {
```
Redi22, line 4: this shouldn't be a class the constants should exist without being an attribute to a class

Redi22 (changes requested):

Redi22 (approved):
LGTM

PR #122, opened 2023-05-03
fix(matador.joshua): add forget password

EyasuTesfu:
fix(matador.joshua): add forget password
- older: forgot_password widget was left out by .gitignore
- fix: forgot_password widget pushed


Redi22 (approved):
LGTM

PR #129, opened 2023-05-04
fix(mobile-dart knights ): merge write article and article reading into one feature article

ben-on:
The two features write article written by Biniyam and article reading written by Samuel are merged into one feature article.


Mobile/starter_project_dark_knights/lib/features/article/presentation/widgets/BookmarkIconButton.dart
Redi22, line 1: rename it following the conventions.

Redi22 (changes requested):

PR #147, opened 2023-05-08
dom[Mobile.dartsmiths]: Add user profile entities and usecase

Zalbo-GJ:
Implemented entities and use cases for User profile. I have done this for the (GetUserProfile) use case and have passed the test.


Mobile/starter_project_dartsmiths/lib/features/profile/domain/usecases/get_user_profile.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:dartz/dartz.dart';
+
+import '../../../../core/error/profile_failures.dart';
+import '../entities/user_profile.dart';
+import '../repositories/user_profile_repository.dart';
+
+class GetUserProfile {
+  final UserProfileRepository repository;
```
lealemgetahun, line 8: The use case class should implement an abstract class use case 
      class GetUserProfile implements UseCase<UserProfile, Params>

PR #150, opened 2023-05-08
feat(backend-dotnetforce7): Implement user authentication and autherization 

brukted:
This PR adds a new project, BlogApp.Identity, which provides user registration and authentication functionality for the application. Users can fetch a list of all users and details of a single user. The BlogApp.Application project is updated to expose the Identity functionality as services, which can be used by the application layer.


Backend/starter_project_dotnetforce7/BlogApp2/BlogApp.API/Controllers/UserController.cs
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        [HttpPost("login")]
+        public async Task<ActionResult<AuthResponse>> Login(AuthRequest request)
+        {
+            return Ok(await _authenticationService.Login(request));
+        }
+
+        [HttpPost("register")]
+        public async Task<ActionResult<RegistrationResponse>> Register(RegistrationRequest request)
```
danielteferadeti, line 25: The same here.

Backend/starter_project_dotnetforce7/BlogApp2/BlogApp.Identity/Services/AuthService.cs
```diff
@@ -0,0 +1,138 @@
...
+            _signInManager = signInManager;
+        }
+
+        public async Task<AuthResponse> Login(AuthRequest request)
+        {
+            var user = await _userManager.FindByEmailAsync(request.Email);
+
+            if (user == null)
+            {
+                throw new Exception($"User with {request.Email} not found.");
```
danielteferadeti, line 39: It would be better to use the result response instance here too (either here or in your controller)

Backend/starter_project_dotnetforce7/BlogApp2/BlogApp.Application/Models/Identity/User.cs
```diff
@@ -0,0 +1,17 @@
+using BlogApp.Domain.Common;
+
+namespace BlogApp.Domain.Models.Identity
+{
+    public enum UserRole
+    {
+        ADMIN,
+        USER
+    }
```
danielteferadeti, line 9: we have userRole defined but am not sure if we have utilized them!

brukted:
@danielteferadeti check it out

danielteferadeti (approved):
This looks great to me! Nice work!

PR #153, opened 2023-05-08
feat(web-iconics): Implement services section of homepage

Henok-Matheas:
services

Implemented the services section of the homepage/


Web/starter_project_iconics/components/home/Services.tsx
```diff
@@ -0,0 +1,22 @@
+import services from '@/data/home/service'
+import React from 'react'
+import IndividualService from './IndividualService'
+
+const Services: React.FC = () => {
+  return (
+    <div className="flex flex-col items-center p-28 gap-y-20">
+      {services.map((service, index) => (
+        <IndividualService
+          key={service.title}
```
Adona12, line 10: It would be better if the key was id was instead of title

Adona12 (approved):

PR #158, opened 2023-05-08
feat(dartsmith.Tamirat).implement update and post entity and usecase with test

tamiratdereje:


tamiratdereje:
I Implemented entity and Usecases for update and post features and also write respective tests for them.

PR #159, opened 2023-05-08
feat(matador.joshua): Add auth domain

EyasuTesfu:
The auth domain consists of:
- User Model with email and password
- User Repository
- authenticate use case


Mobile/starter_project_matador/lib/features/login/Domain/usecases/login_use_case.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:matador/features/login/Domain/repositories/user_repository.dart';
+
+import '../../../../core/failure.dart';
+import '../entities/user.dart';
+import 'package:dartz/dartz.dart';
+
+class GetUser {
+  final UserRepository? repository;
```
lealemgetahun, line 8: implement Usecase class here.

lealemgetahun (approved):

PR #164, opened 2023-05-08
feat(web-iconics): implement impact stories section of homepage

Henok-Matheas:
impact-stories

Implement Impact stories section of the homepage


Web/starter_project_iconics/data/home/impact-story.ts
Adona12, line 1: If it's holding only the data and is inside the data folder, it should be a `.json` file

Adona12 (approved):
LG

PR #180, opened 2023-05-10
fix(backend-dotnetforce7): fix testing structure

gadisamenu:
Changed the  UnitOfWork save method to return 1 and updated the rate test after interlinking with blog.


TigistShiferaw (approved):
Approved. You can merge it. first resolve the conflicting file

TigistShiferaw:
But there is a conflict file so resolve it first

PR #182, opened 2023-05-10
fix(we-barbies) : Fix single blog route

TsehaynehGetaneh:
Fix single blog route 
- fixing single blog route when a user clicks a read more button


ruthgdb (approved):
LGTM. Good job.

PR #183, opened 2023-05-10
feat(Matador):Article-Domain

MuleHakim:
Add article domain which includes entities and usecases, and refactor it


Mobile/starter_project_matador/lib/features/article/domain/usecases/article_usecase.dart
```diff
@@ -0,0 +1,25 @@
...
+
+   GetArticle(this.repository);
+   
+  @override
+  Future<Either<Failure, Article>> call(Params params) async {
+    return await repository.getArticle(params.articleId);
+  } 
+}
+
+class Params extends Equatable{
```
lealemgetahun, line 19: You dont need to use params class to pass in a single parameter

lealemgetahun (changes requested):
Your branch is behind the main so rebase before sending a pr

PR #189, opened 2023-05-10
feat(Backend-zsilentcoders): implementing user authentication

yonass08:
This pull request updates the user authentication process in the backend project by implementing JSON Web Token (JWT) based authentication.

In this update, we have added a new authentication middleware that verifies the JWT token and grants access to the authorized user. The JWT token is generated at the time of user login and is then used to authenticate the user for subsequent requests. The token contains user information and is signed using a secret key to ensure its authenticity.

We use a separate Identity database for the login service which is provided by Identity framework. We used it in conjuction with MailKit to send emails for password reset and confirming emails. And we used google smtp server with our personal account to send the emails.

We also implemented a separate user model so that during integration phase it can easily be integrated with other models, since it is on the same database. And we linked this user to the identity user by including a field called AccountId which refers to the identity user on the identity database.

we will remove the delete and create actions on the user controller and the delete action in the authcontroller, as they are only there for testing purposes and shouldn't be on the production version.

### DIFF
--- a/Backend/starter_project_zsilentcoders/BlogApp3/BlogApp.Application/Features/Users/CQRS/Handlers/Get_UserListQueryHandler.cs
+++ b/Backend/starter_project_zsilentcoders/BlogApp3/BlogApp.Application/Features/Users/CQRS/Handlers/Get_UserListQueryHandler.cs
@@ -25,7 +25,6 @@ namespace BlogApp.Application.Features.Users.CQRS.Handlers
         public async Task<List<_UserDto>> Handle(Get_UserListQuery request, CancellationToken cancellationToken)
         {
             var users = await _unitOfWork._UserRepository.GetAll();
-            Console.Write(users);
             return _mapper.Map<List<_UserDto>>(users);
         }
     }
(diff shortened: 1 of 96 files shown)


danielteferadeti:
Very nice description of your task! Best one I have seen sofar...thank you for that!

Backend/starter_project_zsilentcoders/BlogApp3/BlogApp.Application/Features/Users/CQRS/Handlers/Get_UserListQueryHandler.cs
```diff
@@ -0,0 +1,32 @@
...
+        public Get_UserListQueryHandler(IUnitOfWork unitOfWork, IMapper mapper)
+        {
+            _unitOfWork = unitOfWork;
+            _mapper = mapper;
+        }
+
+        public async Task<List<_UserDto>> Handle(Get_UserListQuery request, CancellationToken cancellationToken)
+        {
+            var users = await _unitOfWork._UserRepository.GetAll();
+            Console.Write(users);
```
danielteferadeti, line 28: This should be removed.

Backend/starter_project_zsilentcoders/BlogApp3/BlogApp.Identity/Configurations/RoleConfiguration.cs
danielteferadeti, line 1: This section look good

Backend/starter_project_zsilentcoders/BlogApp3/BlogApp.Application/Features/Users/CQRS/Handlers/Get_UserListQueryHandler.cs
yonass08: We have removed it

danielteferadeti (approved):
Looks good!

PR #195, opened 2023-05-11
feat(mobile-dartknight):implement Authentication feature usecases

FenetShewarega:
Team : mobile
Team name : dart knight 
Task name implement domain layer for the authentication feature 
Task description: This PR implements authentication use cases in the domain layer..


lealemgetahun (changes requested):
Split it into two PRs one for the UI changes another for the entities and usecases

PR #211, opened 2023-05-12
Profile with state

Segnicho:
added fetching and updating user profile: firstname, lastname, and email.


Web/starter_project_vanguard/package.json
```diff
@@ -15,12 +15,13 @@
     "cy:open": "cypress open"
   },
   "dependencies": {
-    "@reduxjs/toolkit": "^1.9.3",
+    "@reduxjs/toolkit": "^1.9.5",
```
Rediet-Ferew, line 18: there is redundant dependency, it may become a problem when merging

Web/starter_project_vanguard/components/common/Loading.tsx
```diff
@@ -0,0 +1,33 @@
+export const LoadingSpinner = (props: { size?: number }) => {
+    return (
+      <div role="status">
```
jebessaD, line 3: Rather than creating a custom spinner, It is good to use icons like `AiOutlineLoading3Quarters` with the `animate-spin` class. check this out [link](https://tailwindcss.com/docs/animation)

jebessaD (changes requested):
Please change the comment I left for you

Web/starter_project_vanguard/package.json
bisrat-walle, line 18: I don't recommend changing the package.json file as it may affect others code

bisrat-walle (changes requested):

jebessaD (dismissed):
LG

bisrat-walle (approved):

PR #221, opened 2023-05-15
home page domain and data layer

sefinehtesfa34:
Tasks:
home page domain layer
home page data layer
home page domain layer testing
home page data layer testing


lealemgetahun:
Have change to other teams file

PR #224, opened 2023-05-15
feat(backend-dotnetfroces7): Add new user in blog and link user with blog

Biruk-Tassew:
In this pr, we created a new user model the BlogAppDbContext could access and made sure every time a new user is created in the identity user it's also propagated to the new user(BlogUser) table.

We've also created the relation users should have with blog. For the other models we should follow the same steps we followed to create relations.


TigistShiferaw (approved):

TigistShiferaw:
Approved. You can merge it now.

PR #229, opened 2023-05-15
Brook.about us top section

BrookW-83:
Screenshot 2023-05-15 045334


Web/starter_project_iconics/data/about/about-card.json
```diff
@@ -0,0 +1,38 @@
+[
+{
+    "logo": "/img/about/a2sv-program-icons/icon-tag.png",
```
kkstarkk, line 3: since the path to the image file is the same except for the filename just store the image file name and use the name like "/img/about/a2sv-program-icons/{file name goes here}" in the components

kkstarkk (commented):
Looks good..but i have left one minor comment

PR #231, opened 2023-05-15
Brook.card section

BrookW-83:
Screenshot 2023-05-15 053519


Web/starter_project_iconics/styles/globals.css
```diff
@@ -1,7 +1,7 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
-
+/*
```
Mekdeskebede, line 4: better not to use(update) the global css

Web/starter_project_iconics/pages/_app.tsx
```diff
@@ -3,6 +3,7 @@ import Navbar from '@/components/layout/Navbar'
 import '@/styles/globals.css'
 import type { AppProps } from 'next/app'
 
+
```
Mekdeskebede, line 6: remove unnecessary spaces.

Web/starter_project_iconics/components/about/A2svSessionCardItem.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+
+interface props {
```
blenassefa2, line 3: props should be Props since it is interface name the rest LGTM

PR #232, opened 2023-05-15
feat(backend-dotnetforces7): Add user blog relation

Biruk-Tassew:
In this PR, we have created a different user model for the database context blogs and other models are using. This way, we're able to access users securely from both contexts.

We also added the relation user have with blog, and the other models could also be done the same way.


TigistShiferaw (approved):
Approved. You can merge it

PR #281, opened 2023-05-22
(Mobile-Dart-Kinght): Implement data layer for authentication

Feruz2:
Title: Implement Data layer for authentication feature
Description: Added remote and local data layer with test
                    Added repositroy implementation with test


Redi22 (approved):

Feruz2:
Title: Implement data layer
Group: Dart-knight

Description: Added remote and local data layer
                   Added data layer repository
                   Wrote a test for both remote and local data layer
                   Wrote a test for data layer repository

